PR #1, opened 2025-02-23
Refactor

kirilklein:
Streamlined concept processing pipeline


MEDS_preprocess/configs/MEDS.yaml
```diff
@@ -3,54 +3,88 @@ env: azure
...
-filtering: false
+  file_type: csv 
 
 preprocessor:
   _target_: preprocessors.preMEDS.MEDSPreprocessor
 
+data_path:
+  concepts:
+    datastore: researcher_data
+    dump_path: 'data-backup/SP-dumps/2024-09-10'
```
Montgomeryyyy, line 14: this works for now, but will need to be changed when we add the register data

MEDS_preprocess/preprocessors/preMEDS.py
```diff
@@ -1,281 +1,363 @@
...
+        ):
+            processed_chunk = self.concept_processor.process_concept(
+                chunk, concept_config, subject_id_mapping
+            )
+            mode = "w" if first_chunk else "a"
+            self.data_handler.save(processed_chunk, concept_type, mode=mode)
+            first_chunk = False
+
+
+class DataHandler:
```
Montgomeryyyy, line 297: I think it would make sense to move this out of preMEDS and into its own processor, so that is is more clear that it is not related to MEDS, and thus looks cleaner when other functions (like normalisation) uses it

tests/test_preprocessors/test_concept_processor.py
```diff
@@ -0,0 +1,110 @@
+import unittest
+import pandas as pd
+from datetime import timedelta
+from MEDS_preprocess.preprocessors.constants import (
+    ADMISSION, CODE, DISCHARGE, FILENAME,
+    MANDATORY_COLUMNS, SUBJECT_ID, TIMESTAMP
+)
+from MEDS_preprocess.preprocessors.preMEDS import ConceptProcessor
+
+class TestConceptProcessor(unittest.TestCase):
```
Montgomeryyyy, line 10: maybe makes sense to include the script that generates the example data

Montgomeryyyy (changes requested):

MEDS_preprocess/preprocessors/preMEDS.py
kirilklein: I think the restructuring deserves its own PR, see
#2

MEDS_preprocess/configs/MEDS.yaml
kirilklein: yes

tests/test_preprocessors/test_concept_processor.py
kirilklein: Opened #5

PR #6, opened 2025-03-07
Register processing

kirilklein:
## Summary by CodeRabbit

- **New Features**
  - Automated workflows now ensure consistent code formatting, linting, and testing.
  - Enhanced PREMEDS pipeline capabilities with improved backup routines and flexible configurations for both local and Azure environments.
  - Added synthetic data generation tools alongside expanded test suites to validate robust, high-quality performance.

- **Documentation**
  - New comprehensive README for the PREMEDS pipeline detailing structure and functionality.
  - Updated user documentation now provides a comprehensive guide to configuring and using the PREMEDS pipeline.

- **Refactor / Chores**
  - Legacy scripts and outdated settings have been removed to streamline operations and maintain codebase efficiency.


kirilklein:
@coderabbitai review

ehr2meds/PREMEDS/configs/MEDS_azure.yaml
```diff
@@ -0,0 +1,203 @@
...
+      code:
+        column: fill_code
+        regex: '\((D[^)]+)\)'
+    code_prefix: "D/"
+
+  procedure:
+    filename: CPMI_Procedurer.parquet
+    rename_columns:
+      ProcedureCode: code
+      ProcedureName: fill_code
```
Montgomeryyyy, line 52: I don't think it should be the default to fill the procedure codes with the procedure names, as I think that could lead to a messy situation when trying to do mapping/cutoffs, as we would have to read the names. These are also most of the times just danish sentences with the structure as such.

ehr2meds/PREMEDS/configs/MEDS_azure.yaml
```diff
@@ -0,0 +1,203 @@
...
+      timestamp:
+        column: fill_timestamp
+    code_prefix: "M/"
+
+  labtest:
+    filename: CPMI_Labsvar.parquet
+    rename_columns:
+      BestOrd: code
+      Prøvetagningstidspunkt: timestamp
+      Bestillingsdato: fill_timestamp
```
Montgomeryyyy, line 80: I'm also a bit unsure about this one as some values could be ordered months/years in advance if it is some sort of screening, so that would mess with the history. But I think it is more valid, as there is a value in something being ordered. These tests would however, most likely not include values.

ehr2meds/PREMEDS/configs/MEDS_local.yaml
```diff
@@ -0,0 +1,200 @@
...
+      hovedspeciale_ans: clinical_specialty
+      region_ans: region_code
+
+    combine_datetime:
+      timestamp:
+        date_col: date_start
+        time_col: time_start
+        drop_original: true
+
+    unroll_columns:
```
Montgomeryyyy, line 125: It maybe makes sense to add some sort of filtering function as well, as I think most people will have "Ukendt" (unknown) as a value in their reason for contact

ehr2meds/PREMEDS/configs/MEDS_local.yaml
```diff
@@ -0,0 +1,200 @@
...
+        column: fill_timestamp
+    code_prefix: "L/"
+    numeric_columns: [numeric_value]
+
+register_concepts:
+  register_diagnosis:
+    filename: diagnoser.asc
+    rename_columns:
+      diagnosekode: code
+      diagnosetype: code_type
```
Montgomeryyyy, line 91: how is this used?

Montgomeryyyy (changes requested):
I think it looks really good:)
My comments are mostly related to the logic of which things to includes/exclude. 
Additionally, I'm not sure why, but the admissions in the dump in AKK seems to be missing some of the discharges, so there might be bug somewhere relating to this.

ehr2meds/MEDS/MEDS_tabs/task_example.sh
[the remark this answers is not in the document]
kirilklein: Good point

ehr2meds/PREMEDS/configs/MEDS_azure.yaml
kirilklein: Yeah, it's not ideal. But also dropping the latest is not so great. Maybe we just process it twice, creating a labtest, and labtest_ordered csv?

ehr2meds/PREMEDS/configs/MEDS_local.yaml
kirilklein: My idea was to push filtering to create data. But we can also include that here as a feature. Let's make it a separate PR though.

ehr2meds/PREMEDS/configs/MEDS_local.yaml
kirilklein: It isn't. Should be dropped or unrolled as a separate code.

ehr2meds/PREMEDS/configs/MEDS_azure.yaml
kirilklein: Agree.

ehr2meds/PREMEDS/configs/MEDS_local.yaml
Montgomeryyyy: Yeah I think it would make sense to address here, as it is basically a non-valid code like a None, whereas I think the filtering in create data should be more for filtering out valid codes, but no filtering should also just include valid codes

Montgomeryyyy (approved):
looks good

ehr2meds/PREMEDS/preprocessing/premeds/extractor.py
```diff
@@ -0,0 +1,231 @@
...
+            self.cfg.patients_info.filename,
+            cols=list(self.cfg.patients_info.get("rename_columns", {}).keys()),
+        )
+        # Use columns_map to subset and rename the columns.
+        df = select_and_rename_columns(
+            df, self.cfg.patients_info.get("rename_columns", {})
+        )
+        self.logger.info(f"Number of patients after selecting columns: {len(df)}")
+
+        df, hash_to_int_map = factorize_subject_id(df)
```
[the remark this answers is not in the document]
kirilklein: I added +2, so it should be fine now.

PR #33, opened 2026-04-07
The Purge

Sllambias:
Remove deprecated code. If code is still relevant for Azure reasons move it to /azure dir.
Next up will be to streamline pyproject.toml and work on directory structure of remaining code, but that is to be a separate follow-up issue.

I have done a pretty extensive "cleanup" so just let me know what should be spared as I'm sure I've removed code that is still alive.

E.g. @ZahraSobhaninia are you using the Azure logic in this repo?

## Summary by CodeRabbit

* **New Features**
  * Environment-variable driven config/data paths and new CLI commands for PREMEDS (convert/normalize) and synthetic data generation.

* **Documentation**
  * README updated with simpler local workflows; added synthetic data generation guide.

* **Refactor**
  * Reworked preMEDS data-loading and normalization workflows; removed cloud/Azure automation entrypoints.

* **Chores**
  * Bumped Python requirement to 3.11 and revised project dependency declarations.


mikkelfo (commented):
I took Sebastian's advice to look at the new repo rather than the changes -- so I won't be putting comments in the right places
General comments:
1. `readme.MD` in `ehr2meds/synthetic_data_generation` is outdated
2. `ehr2meds/make_synthetic_data.md` should just be renamed to readme as it contains more than just synthetic data stuff
3. Why is `normalize_premeds.py` a standalone script and not just a PREMEDS configuration?

Other comments:
To preface, we might want to save the following comments for a separate PR. In general, this code looks so much like the old BONSAI code (e.g., custom `config.py` and `constants.py` files) and is very hard to follow with naming like `concepts`, `patients_info`(that we only use for subject_id_mapping, which should only be required if IDs are non-ints), overuse of classes, etc. Imo, I think it's due for a complete rewrite, so it follows BONSAI and the synthetic generation code standards.

Montgomeryyyy (approved):

PR #34, opened 2026-04-30
Extend to cover core SDS registries

RH-MikkelWerling:
## Summary
This branch of the repo was tested (and ran without errors) on the DALY-CARE NGC partition, but we need to confirm generalizability. 
Beyond adding functionality, there was a bug on my end with subjects being cast to object and not string. This might be needed on your end, but it broke the pipeline for me.  
The rest of the PR concerns new functionality added more or less exclusively to RegistryConcepts and utils. 

## Bug fixes
- `map_pids_to_ints`: cast subject_id to `str` instead of `object` before mapping, aligning with `factorize_subject_id` which builds the mapping with string keys. Caused all my rows to be dropped. 

### Utils (`utils.py`)
- `replace_values`: string replacement in columns
- `pad_values`: append suffix to column values unless they already contain a specified substring (needed for irregular timestamps)

### RegisterConceptProcessor (`concept_processors.py`)
- Added `time_stamp_dict` parameter for global timestamp alignment 
- Added preprocessing steps: `replace_values`, `normalize_columns`, `apply_value_map`, `pad_values`
- Added `_combine_datetime_from_parts` for sources storing date, hour and minute in different columns
- Added `_get_mapping_table` to make paths for mapping tables fall back to  `ehr2meds/resources/` for static reference files (needed for the SHAK --> region mapping from `ehr2meds/resources/shak_to_region.parquet`)

### Resources
- Added `ehr2meds/resources/shak_to_region.parquet` mapping SHAK hospital codes to Danish regional codes

### Configs
- Added `configs/preMEDS/lymphoma_ngc.yaml` for the NGC lymphoma cohort (which is intended to function as a good initial config for including SDS data generally)
- Added `configs/MEDS/lymphoma_ngc_event.yaml` for the MEDS conversion step (which is intended to function as a good initial config for including SDS data generally)

## Summary by CodeRabbit

* **New Features**
  * Added end-to-end lymphoma NGC support: MEDS event and preMEDS registers-only configurations for demographics, pathology, diagnoses, contacts, admissions, procedures, medications, and laboratory measurements (including numeric values).

* **Improvements**
  * New configurable utilities for value mapping, normalization, replacements and conditional padding.
  * Enhanced datetime/timestamp handling (combine-from-parts, padding/fallback) and improved mapping-table resolution with optional timestamp-aware processing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/FGA-DIKU/ehr2meds/pull/34?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


ehr2meds/preMEDS/utils.py
```diff
@@ -219,25 +224,70 @@ def unroll_columns(df: pd.DataFrame, concept_config: dict) -> List[pd.DataFrame]
...
+
+
+def normalize_columns(df: pd.DataFrame, concept_config: dict) -> pd.DataFrame:
+    """Normalize column values by casting to int then str, getting rid of leading zeros."""
+    for col in concept_config.get("normalize_columns", []):
+        if col in df.columns:
+            df[col] = (
+                pd.to_numeric(df[col], errors="coerce").astype("Int64").astype(str)
+            )
+            df[col] = df[col].replace("<NA>", None)
```
Sllambias, line 252: Is this how PD natively would format missing or is this "<NA>" coming from a specific source? If this is related to a specific dataset this should probably be abstracted out of the code

ehr2meds/preMEDS/concept_processors.py
```diff
@@ -152,3 +189,44 @@ def _combine_datetime_columns(df: pd.DataFrame, concept_config: dict) -> pd.Data
...
+                hour_col = date_time_cols.get("hour_col")
+                minute_col = date_time_cols.get("minute_col")
+
+                if date_col in df.columns:
+                    dt_str = (
+                        df[date_col].dt.strftime("%Y-%m-%d")
+                        if pd.api.types.is_datetime64_any_dtype(df[date_col])
+                        else df[date_col].astype(str)
+                    )
+                    # dt_str = df[date_col].astype(str)
```
Sllambias: should be removed?

Sllambias (approved):
Looks fine for me, just a few very minor comments

ehr2meds/preMEDS/utils.py
RH-MikkelWerling: Are you referring to the "<NA>"-string? That's pandas native format handling; this shouldn't pertain only to a single dataset.

ehr2meds/preMEDS/concept_processors.py
RH-MikkelWerling: Agree - removing

ehr2meds/preMEDS/utils.py
```diff
@@ -219,25 +224,70 @@ def unroll_columns(df: pd.DataFrame, concept_config: dict) -> List[pd.DataFrame]
...
+    return df
+
+
+def replace_values(df: pd.DataFrame, concept_config: dict) -> pd.DataFrame:
+    """Apply string replacement to specified column."""
+    for col, replacements in concept_config.get("replace_values", {}).items():
+        if col in df.columns:
+            for old, new in replacements.items():
+                df[col] = df[col].str.replace(old, new, regex=False)
+    return df
```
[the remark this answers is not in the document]
mikkelfo: Potentially relevant? @RH-MikkelWerling can just open an issue instead of fixing

ehr2meds/preMEDS/utils.py
```diff
@@ -219,25 +224,70 @@ def unroll_columns(df: pd.DataFrame, concept_config: dict) -> List[pd.DataFrame]
...
+
+
+def pad_values(df: pd.DataFrame, concept_config: dict) -> pd.DataFrame:
+    """Append suffix to column values that don't already contain it."""
+    for col, cfg in concept_config.get("pad_values", {}).items():
+        if col in df.columns:
+            suffix = cfg["suffix"]
+            contains = cfg.get("unless_contains", suffix)
+            mask = ~df[col].astype(str).str.contains(contains, regex=False, na=False)
+            df.loc[mask, col] = df.loc[mask, col].astype(str) + suffix
```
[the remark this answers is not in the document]
mikkelfo: Potentially relevant? @RH-MikkelWerling can just open an issue instead of fixing

ehr2meds/preMEDS/utils.py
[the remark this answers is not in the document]
RH-MikkelWerling: I'm fixing this now. I will note however, that this would quite rarely be a problem. The intended functionality is string replacement. It is therefore fair to assume that the column would not be a mixed type but primarily contain strings.

ehr2meds/preMEDS/utils.py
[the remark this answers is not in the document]
RH-MikkelWerling: Yeah, this one is a good catch. Fixing it as well.

configs/preMEDS/lymphoma_ngc.yaml
```diff
@@ -0,0 +1,418 @@
...
+    code_prefix: "RC/"
+
+  SDS_t_adm:
+    filename: SDS_t_adm.parquet               
+    normalize_columns:
+      - c_spec
+      - c_indm
+      - c_kontaars
+    value_map:
+      c_spec:
```
Montgomeryyyy, line 167: where do these mappings come from? If they are from the t_speciale or similar input it might make more sense to just make a function which processes this

Montgomeryyyy (approved):
I have now also tested the code on the fetal data and added the corresponding configs. I also removed the forced pid_link, as it is no longer needed since all subject_ids should now be consistent across data sources. There are still several quirks in the codebase, but I have created separate issues for the ones I have noticed.

configs/preMEDS/lymphoma_ngc.yaml
RH-MikkelWerling: The mapping is from this; https://www.dst.dk/da/Statistik/dokumentation/Times/forebyggelsesregistret/spec
Yes, I agree - we could make a specific function allocated to this. I made the general mapping functionality as something akin to this problem might pop up later. I'm open for making a more slick solution.

PR #35, opened 2026-05-06
Update env

Montgomeryyyy:
Environment file that works local and on ngc as well. 

## Summary by CodeRabbit

* **Chores**
  * Added Python 3.12 compatibility
  * Added new runtime dependencies for configuration management and environment variable handling
  * Updated fetal data processing configuration settings

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -23,4 +23,4 @@ jobs:
     - name: Lint with flake8
       run: |
         # Check syntax errors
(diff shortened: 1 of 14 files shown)


Sllambias (approved):

RH-MikkelWerling (approved):
Note; as we've talked about, it doesn't seem like the repo actually requires Python=3.12 to work, but I agree that the requirements should force this. Our setup (DALY-CARE) doesn't allow for running meds with python 3.12 (yet), but all the code runs. So no problems; it's just a note for future reference.

PR #51, opened 2026-05-28
Removed mandatory patients_info calls

mikkelfo:
EDIT: This PR is potentially not part of #37 -- Might've been previously fixed and this mapping is not related to the `subject_id_mapping`
1. Made `subject_id_mapping` optional per first part of #37 
>For instance, the code base currently enforces a mapping between pseudo_ids/CPR and patient_ids - but everything on SDS or similar is typically already hashed or something akin to this. 
2. Rephrased `patients_info` to `subject_id_mapping` keyword, making `patients_info` an optional (but standard) table that is handled similarly to the others, resolves #40 
>Currently a central patients info table containing the columns gender birthdate deathdate subject_id is a prerequesit when converting data from raw -> preMEDS. A functionallity that creates this central patients info table with the 4 variables, if they exist different tables, is needed.

## Summary by CodeRabbit

* **Chores**
  * Renamed top-level config key: subject → patients_info
  * Added write_file_type: csv and a subject_id_mapping option to load IDs from external tables
  * Reformatted pre-MEDS table config for clearer subject ID mapping

* **Bug Fixes**
  * Added validation to ensure SUBJECT_ID exists and is integer
  * Improved subject ID mapping: reports missing/unknown IDs with a warning and drops unmatched rows

### DIFF
--- a/configs/preMEDS/fetal_SP_synth.yaml
+++ b/configs/preMEDS/fetal_SP_synth.yaml
@@ -17,8 +17,9 @@ paths:
 
 write_file_type: csv
 subject_id_mapping:
-  filename: ${paths.tables_dir}/patient_info.csv
-  subject_id_col: CPR
+  file: ${paths.tables_dir}/patient_info.csv # Table to derive mapping from (can be None)
+  subject_id_col: CPR # Column to use as the subject_id mapping
+  mapping_id_col: # Column to use as the mapping (can be None, then auto-generated)
 
 tables:
   diagnosis:
--- a/configs/preMEDS/fetal_SP_synth.yaml
+++ b/configs/preMEDS/fetal_SP_synth.yaml
@@ -15,7 +15,10 @@ paths:
   tables_dir: ${oc.env:EHR2MEDS_DATA}/raw/fetal_data/SDS_AND_SP_from_population  
   output: ${oc.env:EHR2MEDS_DATA}/preMEDS/fetal_data/SP
 
-write_file_type: csv 
+write_file_type: csv
+subject_id_mapping:
+  filename: ${paths.concepts}/patient_info.csv
+  subject_id_col: CPR
 
 tables:
   diagnosis:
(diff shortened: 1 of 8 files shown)


ehr2meds/preMEDS/extractor.py
```diff
@@ -62,40 +58,37 @@ def __init__(self, cfg):
...
-        # Use columns_map to subset and rename the columns.
-        df = select_and_rename_columns(df, self.cfg.patients_info.get("rename_columns", {}))
-        logger.info(f"Number of patients after selecting columns: {len(df)}")
+        logger.info(f"Number of patients in dataframe: {len(df)}")
+
+        hash_to_int_map = dict(zip(df[SUBJECT_ID], range(len(df))))
 
-        df, hash_to_int_map = factorize_subject_id(df)
         # Save the mapping for reference.
         with open(f"{self.cfg.paths.output}/hash_to_integer_map.pkl", "wb") as f:
```
Sllambias: Not your contribution but I'd avoid .pkl. This should be openable by standard programs and human-readable for convenience

Montgomeryyyy (approved):
Looks good to me, and also works on ngc for me. I just think we need to update the remaining of the configs to the new setup.

Sllambias (commented):
So, aren't we running into problems if a mapping already exists here? As I see it, this does not support re-using previous mappings. Is that correct? My intuition is, that if the population file already contained the mapping, this should be used and not generated. Also the way we generate the mapping `hash_to_int_map = dict(zip(df[SUBJECT_ID], range(len(df))))` isn't this just a cryptic way of extracting the index of a dataframe?

configs/preMEDS/fetal_SP_synth.yaml
```diff
@@ -15,7 +15,10 @@ paths:
   tables_dir: ${oc.env:EHR2MEDS_DATA}/raw/fetal_data/SDS_AND_SP_from_population  
   output: ${oc.env:EHR2MEDS_DATA}/preMEDS/fetal_data/SP
 
-write_file_type: csv 
+write_file_type: csv
+subject_id_mapping:
```
Sllambias, line 19:
```
subject_id_mapping:
    file: path
    ID_col: CPR
    MAP_col: Hash

subject_id_mapping:
    file: path
    ID_col: CPR
    MAP_col: 

subject_id_mapping:
    file: 
    ID_col: 
    MAP_col: 
```

PR #52, opened 2026-05-29
combined concept processors to one

Montgomeryyyy:
Addressing #37 by combining the concept processors to one where we previously has one for SP and another for SDS. 
By doing this I changed the code to use just one directory for the tables, where the exact location can be specified for each filename. @RH-MikkelWerling with this does your lymph config still work on your data?

## Summary by CodeRabbit

* **Improvements**
  * Higher-precision timestamps (fractional seconds) across data extraction for better temporal accuracy.
  * Standardized dataset path handling via environment-driven paths for more consistent configuration.

* **Refactor**
  * Consolidated preprocessing to a unified table-based processing pipeline, simplifying and streamlining data handling.

* **Chores**
  * Removed and replaced legacy configuration entries with updated table-oriented configs.


ehr2meds/preMEDS/extractor.py
```diff
@@ -96,66 +78,38 @@ def format_patients_info(self) -> Dict[str, int]:
         return hash_to_int_map
 
     def format_concepts(self, subject_id_mapping: Dict[str, int]) -> None:
```
mikkelfo: Minor thing, but let's just call it `format_tables`, such that this function matches the config (that uses `tables_dir`)

mikkelfo (approved):
Looks good. The full paths for each table might be slightly annoying, but it's more expressive and flexible, so I think the pros outweigh the cons.

Secondly, while we're fixing some things here, can we rename `concepts` to `tables` such that our configs match our functions' naming? (Don't do it everywhere, just the files we're touching here)

RH-MikkelWerling (approved):
Worked from raw data to full MEDS format. I did have to hack the versioning of the packages a little bit, but everything still worked as intended, it seems. I fixed some minor naming things, which were inconsequential. I did however note that we require dates to be strings in the patient_info - this does not pertain to this PR however, so I made a different issue for it in  #55 .

PR #59, opened 2026-06-25
Updated MEDS packages and appropiate .yaml and scripts

mikkelfo:
EDIT: Updated to `meds-extract==0.6.1`, deleted unused packages and assign version numbers for exisiting. 
~~Updated to `meds-transforms==0.6.7` (installs `meds==0.4.1`) and `meds-extract==0.6.1`.~~ This has made significant changes, primarily to the event configs.

1. @Montgomeryyyy: Newer MEDS requires `numeric_value` to be strictly floats. For the synthetic pipeline, we have string values ("Negative", "Ikke påvist", etc.), so the `default_event.yaml` has been changed from the `numeric_value` column to `lab_value`. Please check if this breaks your normal configs
2. @Montgomeryyyy @RH-MikkelWerling @Katja-Jagd All previous event configs no longer works and should be changed to the new [dftly](https://github.com/mmcdermott/dftly) syntax. You can see `default_event.yaml` for examples or ask me in Slack.

## Summary by CodeRabbit

* **New Features**
  * Updated the MEDS conversion pipeline’s event conversion flow and improved output path/sharding handling.
  * Enhanced default event mappings with template-based code construction and more consistent inline date/time formatting.
* **Chores**
  * Updated the conversion script to invoke the transform pipeline using positional config arguments.
  * Fixed the package version, refreshed contributor metadata, and tightened/adjusted dependency constraints.
  * Removed the preMEDS configuration loader/wrapper.


Montgomeryyyy (approved):
Looks fine to me, and I can run this with my data on NGC. There is the issue of updating the configs, but I am not sure if this is the best way to go #61

configs/MEDS/default_pipeline.yaml
```diff
@@ -1,6 +1,6 @@
-defaults:
-  - _extract
-  - _self_
+# defaults:
```
Sllambias: remove?

pyproject.toml
```diff
@@ -21,19 +23,16 @@ classifiers = [
...
-    "meds==0.3.3",
-    "meds-transforms==0.1.1",
-    "meds_reader",
-    "torch",
+    "pandas>=3.0.3",
+    "numpy>=2.5.0",
+    "pyarrow>=24.0.0",
+    "tqdm>=4.68.3",
+    "meds-transforms==0.6.7",
+    #"meds==0.4.1", # Installed via. meds-transforms
```
Sllambias: remove?

Sllambias (changes requested):
Wont all configs in configs/MEDS need to be updaetd to this format for before they will work after this PR?
If that is the case, then I think they should be updated before the merge. I'll also happily help with adapting them :)

pyproject.toml
mikkelfo: I personally kept it to show our current MEDS compatability. As the libraries are so interlinked, it felt off to just pin meds-transform. I'd argue that it's better to have for visibility, as I think we would check MEDS versions a couple of times in the future

However, I just realized that `meds-extract` installs both `meds` and `meds-transforms`, so I've updated the PR accordingly, but left the compatiblity as a comment

PR #62, opened 2026-06-25
Pre meds cleanup

### LINKED ISSUE #44: When unrolling columns a new concept is creating even it is empty
When unrolling columns the value in that column may be empty resulting in tokens like "RC_REASON/ "

mikkelfo:
Changelog can be seen in commit history. Every commit has a description of how the old functionality is now handled

## Summary by CodeRabbit

* **New Features**
  * Updated MEDS event mappings to use templated, variable-driven codes/timestamps for demographics, diagnoses, procedures, medications, admissions (IN/OUT), and lab tests.
  * Refactored preMEDS configuration to a columns-first mapping style with explicit subject ID mapping and updated runtime settings.
* **Bug Fixes**
  * Standardized date-only vs full timestamp handling and improved null/fallback behavior for gender and medication/lab timing and values.
* **Refactor**
  * Streamlined the preMEDS data pipeline for direct chunked I/O and simplified processing (value mapping and composed fields).


ehr2meds/preMEDS/data_handler.py
```diff
@@ -13,42 +12,44 @@ class DataHandler:
...
-        test_rows: Optional number of rows to load in test mode
     """
 
     def __init__(
         self,
-        output_dir: str,
-        file_type: str,
+        output_dir: Optional[str] = None,
+        write_file_type: str = "parquet",
         chunksize: Optional[int] = None,
```
Sllambias: `chunksize: Optional[int] = 500000,`
then we can remove 27+28

configs/MEDS/lymphoma_ngc_event.yaml
```diff
@@ -2,123 +2,127 @@ subject_id_col: subject_id
...
+    time: set_time($date_start as '%Y-%m-%d', $v_indtime as '%H', $v_indminut as '%M')
+  priority:
+    code: f"RC_PRIO//{$priority}"
+    time: set_time($date_start as '%Y-%m-%d', $v_indtime as '%H', $v_indminut as '%M')
+  clinical_specialty:
+    code: f"RC_SPEC//{$c_spec}"
+    time: set_time($date_start as '%Y-%m-%d', $v_indtime as '%H', $v_indminut as '%M')
+  region_code:
+    code: f"RC_REG//{$c_sgh}"
+    time: set_time($date_start as '%Y-%m-%d', $v_indtime as '%H', $v_indminut as '%M')
```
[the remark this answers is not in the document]
mikkelfo: This is an existing diff between @RH-MikkelWerling and @Montgomeryyyy -- Perhaps we should find a way to streamline this?

configs/preMEDS/fetal_ngc.yaml
```diff
@@ -23,76 +22,67 @@ paths:
...
-      Afsnit: code
-    melt_table:
-      - target_name: timestamp
-        source_cols: [timestamp_in, timestamp_out]
-        prefix_col: code
-        prefix_map: {timestamp_in: IN/, timestamp_out: OUT/}
-    code_prefix: "ADM/"
+      Flyt_ind: 
+      Flyt_ud: 
+      Afsnit:
```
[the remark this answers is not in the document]
mikkelfo: Not critical as this is two seperate data configs, but it's a recurring issue (see the above comment) that separate data configs have varying renaming (@Sllambias idk if we think of this as an issue though)

configs/preMEDS/fetal_SP_synth.yaml
```diff
@@ -79,14 +65,11 @@ tables:
...
     filename: ${paths.tables_dir}/patient_info.csv
-    rename_columns:
+    columns:
       CPR: subject_id
-      Birthdate: birthdate
-      Deathdate: deathdate
-      Gender: gender
+      Birthdate:
+      Deathdate:
+      Gender:
```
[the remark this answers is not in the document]
mikkelfo: This is flavor. I prefer to keep the column names to the original, so that we perserve the exact mapping and matching

ehr2meds/preMEDS/extractor.py
```diff
@@ -19,29 +19,18 @@ class PREMEDSExtractor:
...
 
         # Create data handler for tables
         self.data_handler = DataHandler(
             output_dir=cfg.paths.output,
-            file_type=cfg.write_file_type,
-            chunksize=self.chunksize,
-            test_rows=cfg.get("test_rows", 1_000_000),
-            test=cfg.test,
+            write_file_type=cfg.write_file_type,
+            chunksize=cfg.get("chunksize"),
```
Sllambias: I think it's simpler to just do cfg.chunksize and let it be a required config arg (in combination with previous comment)

ehr2meds/preMEDS/extractor.py
```diff
@@ -76,46 +63,30 @@ def get_subject_id_mapping(self) -> Union[None, Dict[str, int]]:
...
-                self.process_table_chunks(
-                    table_type,
-                    table_config,
-                    subject_id_mapping,
-                    self.time_stamp_dict,
-                )
+                self.process_table_chunks(table_name, table_config, subject_id_mapping)
             except Exception as e:
-                logger.warning(f"Error processing {table_type}: {str(e)}")
+                logger.error(f"Error processing {table_name}: {str(e)}")
```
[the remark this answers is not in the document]
mikkelfo: @Sllambias Previously this was just a warning. I upgrade to an error log, but idk if we should raise an Exception (if data pipeline takes long time and we have a lot of data?)

ehr2meds/preMEDS/normalizer.py
```diff
@@ -16,7 +16,9 @@ def __init__(self, cfg) -> None:
         self.test = cfg.test
         logger.info(f"test {self.test}")
         self.normalization_type = cfg.data["norm_type"]
-        self._init_data_loader()
+        self.data_handler = DataHandler(
+            chunksize=self.cfg.data.chunksize,
```
Sllambias, line 20: Is this getting chunksize from a different place in the cfg where it is already required?

ehr2meds/preMEDS/processors.py
```diff
@@ -65,80 +36,29 @@ def process(
...
+                drop_source=mapping.get("drop_source", False),
+            )
+        return df
+
+    @staticmethod
+    def _get_mapping_table(data_handler, mapping: dict):
         "Find path and relevant columns in either registry or the resources folder."
         filename = mapping["via_file"]
-        cols = [mapping["join_on"], mapping["target_column"]]
+        cols = {mapping["join_on"]: None, mapping["target_column"]: None}
```
Sllambias: Not quite sure I understand this dictionary, it looks a little bit hacky

Sllambias (approved):
Really nice PR, made a lot of the code much more understandable, and also in a size that is much more manageable. I only had two quite minor comments but otherwise all good for me assuming the functionality is tested with at least one of the revamped configs

ehr2meds/preMEDS/normalizer.py
mikkelfo: The `normalizer` code has its own specific config and syntax that doesn't match the preMEDS stuff (see `preMEDS/normalizer.yaml`). It's generally deprecated (Maria is fixing it), so idk if we should streamline this file

ehr2meds/preMEDS/processors.py
mikkelfo: `data_handler.load` requires an argument `cols: Mapping[str, Optional[str]]` for loading in any files. As the `mappings` config has a different setup (requiring additional info than the standard select/rename that `.load` normally does), there's a small mismatch between the two use cases. 

We could streamline `mappings` to include a `columns` arg, but this would require us to duplicate it in the config instead of this one liner (that I admit looks weird)
```
columns:
  join_on_col:
  target_col: 
```

Perhaps changing it to `cols = dict.from_keys([mapping["join_on"], mapping["target_column"]])` is better?

configs/MEDS/lymphoma_ngc_event.yaml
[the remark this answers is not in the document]
RH-MikkelWerling: ah yeah, it should just be mapped to the same prefix and concept. @Montgomeryyyy , we should settle this between ourselves - maybe tomorrow?

ehr2meds/preMEDS/extractor.py
```diff
@@ -19,29 +19,18 @@ class PREMEDSExtractor:
 
     def __init__(self, cfg):
         self.cfg = cfg
-        logger.info(f"test {cfg.test}")
```
Katja-Jagd, line 22: Maybe keep this `test` functionality, as it is usefull for debugging :)

mikkelfo: Re-added `cfg.test` that does a `break` after first chunk in the `extractor`

ehr2meds/preMEDS/processors.py
mikkelfo: Changed to `cols = dict.from_keys([mapping["join_on"], mapping["target_column"]])`

ehr2meds/preMEDS/extractor.py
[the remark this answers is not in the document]
mikkelfo: The newest commit fixes this and raises the Exception to prevent silently dropping tables. This forces the entire config to be valid (@Sllambias I think this is a fine pattern, might be annoying sometimes to fail a full run, but this will prevent "silent" errors)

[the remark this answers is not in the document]
Sllambias: I think it's completely fine, the entire config _should be_ valid and rather find out early in the process

PR #64, opened 2026-07-02
Feature/custom stages

RH-MikkelWerling:
Partly a fix using custom stages for #62 . 

However, this also introduces row_ids as a custom stage. This might be possible to largely ignore, but since it was so small, I included it still. 

It can also be that my very extensive custom stage for fixing time columns is largely solved by the new functionality in main. 

## Summary by CodeRabbit

* **New Features**
  * Added a complete lymphoma data pipeline for converting preMEDS data into MEDS format.
  * Added automatic numeric value binning with configurable or custom bin definitions.
  * Preserved non-numeric laboratory results as text values while improving numeric extraction.
  * Added stable row tracking to support deterministic event ordering and processing.

* **Bug Fixes**
  * Improved timestamp handling across clinical events.
  * Corrected patient, hospital, and admission field mappings.
  * Improved duplicate handling during data processing.


configs/MEDS/lymphoma_ngc_event.yaml
```diff
@@ -16,113 +16,114 @@ patients_info:
...
     code: f"RC_REG//{$c_sgh}"
-    time: set_time($date_start as '%Y-%m-%d', $time_start as '%H:%M')
+    time: set_time($date_start, f"{($v_indtime as int)}:{($v_indminut as int)}" as '%H:%M')
 
 # LPR surgical procedures 
 SDS_t_sksopr:
   SDS_t_sksopr:
     code: f"RPS//{$c_opr}"
-    time: set_time($date_start as '%Y-%m-%d', $time_start as '%H:%M')
+    time: set_time($date_start, f"{(($v_otime as float) as int)}:{(($v_ominut as float) as int)}" as '%H:%M')
```
mikkelfo, line 88: `(($COL as float) as int)` seems a bit odd? Doesn't `as int` alone work? $COL should either be an int or a float if it's a number, so it should only need to be converted into an int

configs/MEDS/lymphoma_ngc_event_no_compose.yaml
```diff
@@ -16,113 +16,114 @@ patients_info:
...
 # NOTE: numeric_value is NOT normalized here - needs to adhere to splits to not have information leakage
 SDS_laboratorieproevesvar:
   SDS_laboratorieproevesvar:
-    code: f"L//{$code}"
-    time: set_time($samplingdate as '%Y-%m-%d', $samplingtime as '%H:%M:%S')
-    numeric_value: $samplevalue
\ No newline at end of file
+    code: f"L//{$analysiscode}"
+    time: set_time($samplingdate as '%Y-%m-%d', $samplingtime as '%H:%M:%S%.f')
+    numeric_value: '(extract group 1 of /^\s*([-+]?(?:\d+(?:\.\d*)?|\.\d+)(?:[eE][-+]?\d+)?)\s*$/ from $samplevalue) as float64'
```
mikkelfo: God I hate regex. Perhaps add a comment to make it less voodoo

configs/MEDS/lymphoma_ngc_event_no_compose_with_row_idx.yaml
mikkelfo, line 1: We weren't able to add the row_index column on the fly right? (Due to the event config then saying "dont load in this thing"). If not,  I just vote for having either with or without row index rather than both, so we don't have near duplicates. 

But I would reaaaally love if we could find a solution where we dont have to define it as a column. It seems a bit tedious.

configs/MEDS/lymphoma_pipeline.yaml
```diff
@@ -0,0 +1,73 @@
...
+                      0.91, 0.92, 0.93, 0.94, 0.95, 0.96, 0.97, 0.98, 0.99]
+  - bin_numeric_values_fast:
+      bin_with_columns: ["values/quantiles"]
+      code_with_bin_name: "{code}//value_[{left},{right})"
+      drop_numeric_value: false
+      custom_bins_filepath: null
+  - finalize_MEDS_metadata
+  - finalize_MEDS_data
+
+# stages:
```
mikkelfo: Should prolly deleted this

configs/preMEDS/lymphoma_ngc.yaml
```diff
@@ -18,14 +18,14 @@ paths:
 
 write_file_type: parquet 
 
-subject_id_mapping:
-  file: ${paths.tables_dir}/CPMI_PatientInfo.parquet # Table to derive mapping from (can be None)
-  subject_id_col: patientid # Column to use as the subject_id mapping
-  mapping_id_col: # Column to use as the mapping (can be None, then auto-generated)
+#subject_id_mapping:
```
mikkelfo: Not used anymore?

configs/preMEDS/lymphoma_ngc.yaml
```diff
@@ -128,10 +128,10 @@ tables:
       c_spec: 
       c_sgh: 
       k_recnum: contact_id
-    compose:
-      time_start:
-        columns: [v_indtime, v_indminut]
-        separator: ":"
+    # compose:
```
mikkelfo: should delete if not needed

configs/preMEDS/lymphoma_ngc.yaml
```diff
@@ -211,10 +211,10 @@ tables:
         target_column: patientid
         rename_to: subject_id
         drop_source: true
-    compose:
-      time_start:
-        columns: [v_otime, v_ominut]
-        separator: ":"
+    # compose:
```
mikkelfo: should delete if not needed

configs/preMEDS/lymphoma_ngc.yaml
```diff
@@ -232,10 +232,10 @@ tables:
...
```
mikkelfo: should delete if not needed

ehr2meds/meds_stages/bin_numeric_values_fast.py
mikkelfo, line 1: Is a lot of this copied pasted from MEDS-transform? If so, include a reference to the file+line as a comment

mikkelfo (commented):
Minor things, but generally looks fine. If nobody needs `compose_columns` (which I believe not), lets jsut delete it in this PR

configs/MEDS/lymphoma_ngc_event_no_compose.yaml
RH-MikkelWerling: yeah, it's black magic - I'll add a comment so that it's less witch doctor ish

configs/MEDS/lymphoma_ngc_event.yaml
RH-MikkelWerling: Some of my values had "10.0", which didn't pass. This seemed like the lesser evil.

configs/MEDS/lymphoma_ngc_event_no_compose_with_row_idx.yaml
RH-MikkelWerling: yeah, I wholeheartedly agree - I'll look into this tomorrow.

configs/MEDS/lymphoma_pipeline.yaml
RH-MikkelWerling: agree

configs/preMEDS/lymphoma_ngc.yaml
RH-MikkelWerling: not used by me at least, I'll delete

configs/preMEDS/lymphoma_ngc.yaml
RH-MikkelWerling: agree

configs/preMEDS/lymphoma_ngc.yaml
RH-MikkelWerling: agree

configs/preMEDS/lymphoma_ngc.yaml
RH-MikkelWerling: agree

ehr2meds/meds_stages/bin_numeric_values_fast.py
RH-MikkelWerling: yeah, I'll make it more clear what's new and what's taken directly. Good point!

ehr2meds/meds_stages/add_source_row_id.py
```diff
@@ -0,0 +1,81 @@
...
+def main(cfg: DictConfig) -> None:
+    """Add source-row identifiers before MEDS extraction and sharding."""
+
+    input_dir = Path(str(cfg.stage_cfg.data_input_dir))
+    output_dir = Path(str(cfg.stage_cfg.output_dir))
+
+    if not input_dir.exists():
+        raise FileNotFoundError(f"Input directory does not exist: {input_dir}")
+
+    input_files = sorted(input_dir.rglob("*.parquet"))
```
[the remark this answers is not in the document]
mikkelfo: Ignore this

configs/MEDS/lymphoma_ngc_event.yaml
mikkelfo: Yeah, I guess it's fine then. Nice to have a mix between floats, strings, and ints in a single column...

ehr2meds/meds_stages/bin_numeric_values_fast.py
```diff
@@ -0,0 +1,147 @@
...
+        )
+        .with_columns(
+            pl.col("__edge").cast(pl.String).alias("__left"),
+            pl.when(pl.col("__idx") == pl.col("__n"))
+            .then(pl.lit("inf"))
+            .otherwise(pl.col("__next").cast(pl.String))
+            .alias("__right"),
+        )
+        .select(*join_cols, "__edge", "__idx", "__left", "__right")
+    )
```
Sllambias, line 70: lord have mercy with the soul that needs to change this

RH-MikkelWerling: Yeah, it is really not pretty. There's a balance between staying close to the MEDS_transforms version, and creating something that is a bit easier on the eyes. By popular demand, I'll make this more readable.

mikkelfo: I think just pointing to MEDS-transforms and highlighting the new changes is the way to go. This way, we're not suddenly maintainers of additional code and can rely on MEDS-transform upgrades

Sllambias: Yea sorry didn't mean to send you on extra work with my comment. MEDS is not your fault

RH-MikkelWerling: aaaaaahh just made it way more readable hahaha. 
The thing is that the way they join in MEDS_transforms native implementation is problematic and explodes with regards to memory, so that's the main thing that needs to be resolved. 
But given that we need to solve that problem,  we would have to maintain some sort of version of a custom stage for now.  
We could try then to push for MEDS-transform to update their implementation. It is made deliberately to reproduce their outcomes 1-1. So maybe we should just make a PR for Matthew instead? 
If you are okay with it, let me just finish making the code more readable - I think also passing this on to @Montgomeryyyy at some point will require it to be at least slightly readable.

mikkelfo: Yeah maybe that's fine then, also for Maria's sake. If you make it more readable + efficient, we can always ask Matthew about getting it into MEDS-transforms

RH-MikkelWerling: Great, opting for that. It will be longer due to readability, but it's the same thing.

RH-MikkelWerling: @Sllambias tried to make it much more friendly and less black magic now. I think this makes way more sense, and is much easier to understand / adapt from. Let me know if you agree.

Sllambias: Yea I think it's gone from incomprehensible to actually maintainable code 🥇

RH-MikkelWerling: yeah, thanks for calling this out!

configs/MEDS/lymphoma_ngc_event_no_compose_with_row_idx.yaml
RH-MikkelWerling: Looked into this and unfortunately I don't think there's a better solution currently. I think if we want to keep it as a column, we need to preserve it everywhere in the event configs.

configs/MEDS/lymphoma_ngc_event.yaml
mikkelfo, line 1: Maybe drop the renaming if `compose_columns` is gone going forward

configs/MEDS/lymphoma_pipeline.yaml
```diff
@@ -0,0 +1,64 @@
...
+          tuning: 0.1
+          held_out: 0.1
+  - convert_to_subject_sharded
+  - convert_to_MEDS_events:
+      do_dedup_text_and_numeric: true
+  - merge_to_MEDS_cohort:
+      # ordering primarily on timestamp and patientid, but also on row_idx to ensure deterministic ordering for events with the same timestamp and patientid.
+      # We add source_block to the ordering to ensure that events from different source blocks are ordered consistently, even if they have the same timestamp and patientid. 
+      # Note that ordering for code is given in the event configs explicitly and therefore is not given here.
+      additional_sort_by: [source_block, row_idx]
```
mikkelfo: I think sorting by `source_block` is incorrect

configs/preMEDS/lymphoma_ngc.yaml
mikkelfo: still need to delete

ehr2meds/preMEDS/utils.py
```diff
@@ -1,11 +1,21 @@
 import pandas as pd
 from ehr2meds.preMEDS.constants import (
     MANDATORY_COLUMNS,
+    ROW_INDEX,
     SUBJECT_ID,
 )
 from typing import Dict
 
 
+def add_row_idx(df: pd.DataFrame, start: int = 0) -> pd.DataFrame:
```
mikkelfo, line 10: Hmm, for this is the following needed?

1. Check for row_index. Will this ever happen? And if it will, does it hurt that we overwrite it?
2. Doing a copy? Despite being shallow, I dont see the need (is it to avoid pandas stupid errors?)
3. In-place insert instead of just setting `result[ROW_INDEX] = range(len(df))`

pyproject.toml
```diff
@@ -31,6 +31,7 @@ dependencies = [
     "python-dotenv>=1.0.0",
     "hydra-core>=1.3.2",
     "omegaconf>=2.3.0",
+    "PyYAML>=6.0",
```
mikkelfo: I think we should stick to `omegaconf` instead since we're already using it and it is also what MEDS-extract uses by standard. Then we just keep their dependencies

ehr2meds/meds_stages/augment_event_config.py
mikkelfo, line 1: Could we do this with `from omegaconf import DictConfig, OmegaConf` instead? Then we keep to native MEDS libraries

mikkelfo (changes requested):
I think there's some smaller (and annoying) things we maybe should fix first 🫠

ehr2meds/meds_stages/augment_event_config.py
RH-MikkelWerling: Yes - fixed in the most recent commits.

pyproject.toml
RH-MikkelWerling: agree, fixed in the most recent commit

ehr2meds/preMEDS/utils.py
RH-MikkelWerling: made a simpler approach. Still checking for the column existing though - I guess we don't really need to. I can remove the check.

configs/MEDS/lymphoma_pipeline.yaml
RH-MikkelWerling: absolutely correct - only sorting by row_idx now

configs/MEDS/lymphoma_ngc_event.yaml
RH-MikkelWerling: trie

ehr2meds/preMEDS/utils.py
RH-MikkelWerling: deleted check now

mikkelfo (approved):
Good ol' "lgtm"

PR #65, opened 2026-07-07
Added a small DST pipeline in parquet format for testing

mikkelfo:
Added a small DST pipeline in parquet format to test the found bug (https://github.com/mmcdermott/MEDS_extract/issues/120) 

## Summary by CodeRabbit

* **New Features**
  * Added Parquet support for synthetic data generation and conversion.
  * Introduced a new pipeline test workflow that runs the raw-data-to-MEDS process on pushes and pull requests.
  * Added DST Parquet configuration and updated event mappings for new output fields.

* **Bug Fixes**
  * Improved file handling so output format can be selected more consistently across data generation steps.

* **Chores**
  * Updated a project dependency to the latest patch version.

### DIFF
--- a/.github/workflows/pipeline_parquet.yml
+++ b/.github/workflows/pipeline_parquet.yml
@@ -1,0 +1,49 @@
+name: 'Pipeline parquet test'
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+      
+permissions:
+  contents: read
+  pull-requests: read
+
+env:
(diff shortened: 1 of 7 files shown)


RH-MikkelWerling (approved):
Looks good - thanks for taking charge on this. The only thing is that it defaults to csv when nothing is specified. And while that is probably the most used filetype, parquet would be a faster and better default imo. But it's quite a pedantic point. Thoughts?

PR #66, opened 2026-07-14
Add PHAIR Light configs and timezone fix

Katja-Jagd:
## Summary

This small PR adds configs for the PHAIR Light dataset and includes a small preprocessing fix for timezone-aware datetime columns.

### Changes

- Added preMEDS config for the PHAIR Light dataset.
- Added MEDS config for the PHAIR Light dataset.
- Added a preprocessing step that converts timezone-aware datetime columns to timezone-naive UTC before writing preMEDS outputs. This avoids schema mismatches (`datetime[μs, UTC]` vs. `datetime[μs]`) during the MEDS_extract merge stage while leaving all other column types unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added MEDS and preMEDS configuration support for the `phair_light` data mapping workflow.
  * Added mappings for patient birthdates, diagnoses, procedures, psychiatric records, and related clinical tables.
  * Added configurable output tables with standardized subject, date, and time fields.

* **Bug Fixes**
  * Improved date and time handling by consistently converting timezone-aware values to UTC without timezone metadata.


configs/MEDS/phair_light.yaml
```diff
@@ -0,0 +1,61 @@
...
+    time: $birthdate
+
+epikur:
+  epikur:
+    code: f"PM//{$atc}"
+    time: $eksd
+
+indberetningmedpris:
+  indberetningmedpris:
+    code: f"AM//{$c_atc}"
```
mikkelfo: Using "PM" and "AM" (even though it's prescription vs. administered) will duplicate all our ATC codes. Likely not a good idea unless we have a specific reason for it (we prefer to re-use codes as much as possible)

configs/preMEDS/phair_light.yaml
```diff
@@ -0,0 +1,284 @@
...
+
+paths:
+  output: ${oc.env:EHR2MEDS_DATA}/preMEDS/PHAIR_light
+  tables_dir: /storage/archive/UCPH/PHAIR/dhr/tables/parquet
+  #file_type: parquet
+
+write_file_type: parquet
+
+subject_id_mapping:
+  file: /users/people/katjag_r/person_id_map.parquet
```
mikkelfo: Hardcoded -- is there no file we can use for this for PHAIR? Should be a "befolkning"-ish one I believe

configs/preMEDS/phair_light.yaml
mikkelfo, line 1: This makes it very obvious that we should fix #39 soon as this is used extensively here

ehr2meds/preMEDS/processors.py
```diff
@@ -65,3 +66,11 @@ def _get_mapping_table(data_handler, mapping: dict):
             filename = str(Path(__file__).parent.parent / "resources" / filename)  # TODO: Seems very hacky
 
         return data_handler.load(filename, cols=cols)
+
+    @staticmethod
+    def _remove_timezones(df: pd.DataFrame) -> pd.DataFrame:
```
mikkelfo: This should be placed in utils imo

ehr2meds/preMEDS/processors.py
mikkelfo: Should update the doc string to include the removal of time zones

mikkelfo (approved):
Very minor things, otherwise looks good. #39 needs to be prioritized as PHAIR_light appears to use this functionality extensively

configs/MEDS/phair_light.yaml
Katja-Jagd: Done

ehr2meds/preMEDS/processors.py
Katja-Jagd: Done

ehr2meds/preMEDS/processors.py
Katja-Jagd: Done

configs/preMEDS/phair_light.yaml
Katja-Jagd: Done, changed it to `t_person.parquet` with `v_pnr` as id column

PR #67, opened 2026-07-20
Feature/numeric values

RH-MikkelWerling:
An attempt to implement some of the needed functionality for including numeric values for later use in BONSAI. 

Importantly, it:

- Adds aggregate_numeric_metadata, fitted on training subjects only.
- Computes per-code:
  - training count;
  - p1 and p99 normalization bounds;
  - adaptive requested and effective bin counts;
  - deduplicated quantile edges;
  - bin representatives;
  - constant-value status;
  - configuration version.

- Adds annotate_numeric_values, which preserves the original code and raw numeric_value while producing:
  - numeric_value_normalized;
  - numeric_value_bin;
  - numeric_value_binned;
  - numeric_value_present;
  - numeric_value_was_clipped.

Supports applying frozen numeric metadata from another dataset through numeric_metadata_filepath.

## Summary by CodeRabbit

- **New Features**
  - Added a complete lymphoma MEDS pipeline configuration for longitudinal-to-flattened conversion.
  - Introduced new MEDS transform stages for aggregating numeric metadata, annotating numeric values (normalized/bin outputs), joining bin codes into lab identifiers, and augmenting event configs with shared columns.
  - Exposed these stages via MEDS pipeline entry points and updated the custom stage documentation.

- **Bug Fixes**
  - Improved event timestamp construction for LPR/LPR3 diagnoses, procedures, and contacts with correct start-time derivation and fractional-second precision.
  - Updated LPR procedure handling to align procedure type and time precision.

- **Documentation**
  - Removed the documented preMEDS normalization config/CLI and updated workflow notes to reflect the new MEDS stage-based approach.


configs/MEDS/lymphoma_pipeline.yaml
```diff
@@ -27,37 +27,42 @@ stages:
...
   - convert_to_MEDS_events:
       do_dedup_text_and_numeric: true
   - merge_to_MEDS_cohort:
-      # ordering primarily on timestamp and patientid, but also on row_idx to ensure deterministic ordering for events with the same timestamp and patientid.
-      # Note that ordering for code is given in the event configs explicitly and therefore is not given here.
+      # row_idx provides deterministic ordering for simultaneous events.
       additional_sort_by: [row_idx]
   - extract_code_metadata
-  - aggregate_code_metadata:
+  - aggregate_numeric_metadata:
```
Montgomeryyyy, line 45: since we have added a lot of options based on verbal discussions, I think it would be nice to write this documentation in the read me file explaining the functionality of each stage, and whether or not it is optional for the stages below.

ehr2meds/preMEDS/extractor.py
```diff
@@ -79,17 +79,24 @@ def process_table_chunks(
...
+            input_row_count = len(chunk)
             processed_chunk = self.processor.process(
                 chunk,
                 table_config,
                 self.data_handler,
                 subject_id_mapping,
+                row_index_start=next_row_idx,
             )
 
+            next_row_idx += input_row_count
```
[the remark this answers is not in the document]
RH-MikkelWerling: This I think we have established is not so much a problem.

Montgomeryyyy (commented):
Overall I think it looks really good!
I'll just try to update the configs and run this on the my data + the local data, to test if I run into any issues.

ehr2meds/meds_stages/annotate_numeric_values.py
```diff
@@ -0,0 +1,192 @@
...
+CODE = DataSchema.code_name
+VALUE = DataSchema.numeric_value_name
+NORMALIZED_VALUE = "numeric_value_normalized"
+BIN_INDEX = "numeric_value_bin"
+BINNED_VALUE = "numeric_value_binned"
+DERIVED_COLUMNS = {
+    NORMALIZED_VALUE: pl.Float32,
+    BIN_INDEX: pl.Int32,
+    BINNED_VALUE: pl.Float32,
+}
```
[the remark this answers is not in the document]
RH-MikkelWerling: This is by design. It was in the previous version of this PR, but I think it would almost never be used.

configs/MEDS/lymphoma_ngc_event.yaml
```diff
@@ -16,113 +16,120 @@ patients_info:
...
 
 # LPR3 other procedures
 SDS_procedurer_andre:
   SDS_procedurer_andre:
     code: f"RPO//{$procedurekode}"
-    time: set_time($date_start as '%Y-%m-%d', $time_start as '%H:%M:%S')
+    time: set_time($date_start, $time_start as '%H:%M:%S%.f')
+  procedure_type:
+    code: f"RPO_TYPE//{$proceduretype}" # NOTE: default to "P" if missing, as this is LPR3 behavior
+    time: set_time($date_start, $time_start as '%H:%M:%S%.f')
```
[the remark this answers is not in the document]
RH-MikkelWerling: Ah, this should be edited. The idea is that this is missing in LPR but not in LPR3, so there should be no comment here.

configs/MEDS/lymphoma_pipeline.yaml
RH-MikkelWerling: Total agreement. I think in general, the README needs an update.

ehr2meds/meds_stages/aggregate_numeric_metadata.py
```diff
@@ -0,0 +1,238 @@
...
+import polars as pl
+from meds import DataSchema
+from MEDS_transforms.stages import Stage
+from omegaconf import DictConfig
+
+CODE = DataSchema.code_name
+VALUE = DataSchema.numeric_value_name
+TIME = DataSchema.time_name
+TRAINING_VALUES = "numeric_values/training_values"
+LOWER_BOUND = "numeric/p1"
```
mikkelfo: These match percentiles, but we don't necessary have percentiles. Should probably be `bin_1` or similar (same with p99)

ehr2meds/meds_stages/aggregate_numeric_metadata.py
```diff
@@ -0,0 +1,238 @@
...
+    upper_quantile: float = 0.99
+
+    def __post_init__(self) -> None:
+        if self.min_bins < 1 or self.max_bins < self.min_bins:
+            raise ValueError("bin bounds must satisfy 1 <= min_bins <= max_bins")
+        if not 0 <= self.lower_quantile < self.upper_quantile <= 1:
+            raise ValueError("quantiles must satisfy 0 <= lower < upper <= 1")
+
+    @classmethod
+    def from_stage_config(cls, stage_cfg: DictConfig) -> "NumericBinningConfig":
```
mikkelfo: I don't see why we can't just use the dataclass init instead of a custom `from_stage_config` that does the same

ehr2meds/meds_stages/join_numeric_bins.py
```diff
@@ -0,0 +1,42 @@
...
+        pl.col(NUMERIC_BIN).cast(pl.String),
+    )
+
+    return data.with_columns(pl.when(pl.col(NUMERIC_BIN).is_not_null()).then(joined_code).otherwise(pl.col(CODE)).alias(CODE))
+
+
+def join_numeric_bins_fntr(stage_cfg: DictConfig) -> Callable[[pl.LazyFrame], pl.LazyFrame]:
+    """Build the final code-representation transform."""
+    separator = str(stage_cfg.get("separator", "//"))
+    if not separator:
```
mikkelfo: A little weird pattern with a `.get(key, DEFAULT)` followed by a None check. This only triggers if someone explicitly sets `separator=None`

ehr2meds/meds_stages/annotate_numeric_values.py
mikkelfo, line 1: Heavy file. Idk if we can do anything about it (?)

ehr2meds/meds_stages/aggregate_numeric_metadata.py
mikkelfo, line 1: Heavy file. Idk if we can do anything about it (?)

ehr2meds/meds_stages/join_numeric_bins.py
RH-MikkelWerling: The idea is not to prevent None but empty strings. That could be made more explicit. I'm removing the check because that would still technically work although be a bit bonkers.

mikkelfo: Ah fair. Idk how we could handle that tbh. Maybe this is fine or we just force the `//`, likely doesnt matter much

RH-MikkelWerling: I think if anything we should be more verbose, because I agree that your reading was more natural. Like either we should explicitly check for empty strings after handling nulls or else we should just not check. 
I think not checking is also fine actually, but it's sort of a design choice how much handholding we want.

ehr2meds/meds_stages/aggregate_numeric_metadata.py
RH-MikkelWerling: I think it can be shortened a bit - but not a lot if we also want it to be readable. 
One thing we could do is split this into two stages; one for just making all the metadata and one for aggregating. But that's the best I've got right now.

ehr2meds/meds_stages/join_numeric_bins.py
mikkelfo: Since it's not critical if a sep is "" (just leads to weird codes) and that `//` is forced by default, perhaps we can just drop the check?

ehr2meds/meds_stages/aggregate_numeric_metadata.py
mikkelfo: Honestly idk. @Sllambias Maybe you have a good idea?

ehr2meds/meds_stages/join_numeric_bins.py
RH-MikkelWerling: exactly - I'm pushing in a minute when I've looked at whether we can reduce the big files.

ehr2meds/meds_stages/aggregate_numeric_metadata.py
RH-MikkelWerling: It could make sense; I just don't know of a case when we would want metadata without using it downstream. But for ease of reading, it would be nicer. But I don't know how much of an argument that is in itself. Let me know what you think.

mikkelfo: I think keeping them together _if_ they are always going to be called together is better

RH-MikkelWerling: Agree. Can you think of a situation, where we want to calculate metadata but not run the aggregation or vice versa? Even when external data is provided, the current implementation fills out the blanks if there are codes that were not included in the external metadata.

ehr2meds/meds_stages/annotate_numeric_values.py
RH-MikkelWerling: I tried, but I'm unsure that you can do that much about it. Maybe fresher eyes can actually reduce it.

ehr2meds/meds_stages/aggregate_numeric_metadata.py
RH-MikkelWerling: True, resolved.

ehr2meds/meds_stages/aggregate_numeric_metadata.py
RH-MikkelWerling: true, resolved

ehr2meds/meds_stages/aggregate_numeric_metadata.py
```diff
@@ -0,0 +1,211 @@
...
+    def __post_init__(self) -> None:
+        if self.min_bins < 1 or self.max_bins < self.min_bins:
+            raise ValueError("bin bounds must satisfy 1 <= min_bins <= max_bins")
+        if not 0 <= self.lower_quantile < self.upper_quantile <= 1:
+            raise ValueError("quantiles must satisfy 0 <= lower < upper <= 1")
+
+
+def calculate_bin_count(n_unique: int, minimum: int, maximum: int) -> int:
+    """Return the bounded adaptive bin count B(N)=1.14*N_unique**0.237."""
+    if minimum < 1 or maximum < minimum:
```
mikkelfo: I believe this is already checked through the `NumericBinningConfig`, as the `minimum, maximum` is set by the config. No need to double check it

RH-MikkelWerling: ah true, the post init checks it

ehr2meds/meds_stages/aggregate_numeric_metadata.py
```diff
@@ -0,0 +1,238 @@
...
+
+
+def _fit_transform(values: list[float], config: NumericBinningConfig) -> dict[str, object]:
+    """Transform and annotate one code."""
+    series = pl.Series(sorted(values), dtype=pl.Float64)
+    lower = float(series.quantile(config.lower_quantile, interpolation="linear"))
+    upper = float(series.quantile(config.upper_quantile, interpolation="linear"))
+    n_bins = calculate_bin_count(series.n_unique(), config.min_bins, config.max_bins)
+
+    if upper <= lower:
```
Sllambias: Should this ever be the case? I'm afraid we're letting things be valid for the sake of avoiding code crashes

ehr2meds/meds_stages/aggregate_numeric_metadata.py
```diff
@@ -0,0 +1,238 @@
...
+    max_bins: int = 100,
+    lower_quantile: float = 0.01,
+    upper_quantile: float = 0.99,
+) -> pl.DataFrame:
+    """Combine training shards and make per-code transformations.
+
+    The intended behavior is to receive only training shards, configured
+    by running the command with ``train_only: true``.
+    """
+    config = NumericBinningConfig(
```
Sllambias: Feels like the concept of a config can be entirely written out

ehr2meds/meds_stages/aggregate_numeric_metadata.py
```diff
@@ -0,0 +1,238 @@
...
+TRAINING_VALUES = "numeric_values/training_values"
+LOWER_BOUND = "numeric/p1"
+UPPER_BOUND = "numeric/p99"
+BIN_EDGES = "numeric/bin_edges"
+BIN_REPRESENTATIVES = "numeric/bin_representatives"
+HARD_MINIMUM = "numeric/hard_minimum"
+HARD_MAXIMUM = "numeric/hard_maximum"
+
+
+@dataclass(frozen=True)
```
Sllambias: I would remove this class and just use `x=min_bins` rather than `x=config.min_bins`

ehr2meds/meds_stages/aggregate_numeric_metadata.py
```diff
@@ -0,0 +1,238 @@
...
+from dataclasses import dataclass
+from datetime import date, datetime, time
+from pathlib import Path
+
+import polars as pl
+from meds import DataSchema
+from MEDS_transforms.stages import Stage
+from omegaconf import DictConfig
+
+CODE = DataSchema.code_name
```
Sllambias: I think this is a very confusing pattern that we're defining hardcoded config arguments in python files and it becomes even more confusing when we define them in one python file to be used in another and even even more when we define the same variable names in different python files that also import from each other. 
All of these should be extracted to 1 _actual_ config file I think

Sllambias: And I'm aware this is not your pattern and that some of these may be unavoidable, but some of them we can avoid I think.
And for the unavoidable MEDS-variables I personally prefer `fn(x=DataSchema.code_name)` rather than 
```
y = DataSchema.code_name
fn(x=y)
```
as it is much more explicit, but I know that this is a matter of taste

ehr2meds/meds_stages/annotate_numeric_values.py
```diff
@@ -0,0 +1,192 @@
...
+from meds import CodeMetadataSchema, DataSchema
+from MEDS_transforms.stages import Stage
+from MEDS_transforms.utils import PKG_PFX, resolve_pkg_path
+from omegaconf import DictConfig
+from pathlib import Path
+
+CODE = DataSchema.code_name
+VALUE = DataSchema.numeric_value_name
+NORMALIZED_VALUE = "numeric_value_normalized"
+BIN_INDEX = "numeric_value_bin"
```
Sllambias: Same comment here regarding the python / yaml config files

ehr2meds/meds_stages/annotate_numeric_values.py
Sllambias: I think the pattern of functions using locally hardcoded variables and being called as "fn()" is a very implicit structure. `usable_value()` is an example of something completely opaque, it could literally check anything and noone knows what value is checked whereas `usable(value=x, hard_max=max_val, hard_min=min_val, lower_bound=bound` at least gives some context and traceability.

Sllambias (commented):
Overall I have one comment, which I'm aware is also a personal taste, but I think that prefixing functions with underscores just feels like noise, so I'd prefer to not use those and then a few file-specific comments on hardcoded variables and config-styles

ehr2meds/meds_stages/aggregate_numeric_metadata.py
```diff
@@ -0,0 +1,216 @@
...
+import math
+import polars as pl
+from collections.abc import Callable
+from datetime import datetime
+from meds import DataSchema
+from MEDS_transforms.stages import Stage
+from omegaconf import DictConfig
+from pathlib import Path
+
+TRAINING_VALUES = "numeric_values/training_values"
```
Sllambias: See review comment 1

ehr2meds/meds_stages/annotate_numeric_values.py
```diff
@@ -0,0 +1,197 @@
...
+
+    missing_bounds = [column for column in BOUND_COLUMNS if column not in metadata.columns]
+    metadata = metadata.with_columns(*(pl.lit(None, dtype=pl.Float64).alias(column) for column in missing_bounds))
+    return metadata.select(*key, *METADATA_COLUMNS)
+
+
+def combine_numeric_metadata(
+    fitted_metadata: pl.DataFrame,
+    external_metadata: pl.DataFrame,
+    *,
```
Sllambias: See review comment 3

ehr2meds/meds_stages/aggregate_numeric_metadata.py
```diff
@@ -0,0 +1,216 @@
...
+HARD_MAXIMUM = "numeric/hard_maximum"
+
+
+def calculate_bin_count(n_unique: int, minimum: int, maximum: int) -> int:
+    """Return the bounded adaptive bin count B(N)=1.14*N_unique**0.237."""
+    estimated = round(1.14 * n_unique**0.237)
+    return max(minimum, min(maximum, estimated))
+
+
+def read_value_bounds(stage_cfg: DictConfig) -> pl.DataFrame | None:
```
Sllambias: See review comment 2

ehr2meds/meds_stages/aggregate_numeric_metadata.py
```diff
@@ -0,0 +1,216 @@
...
+
+    above_minimum = pl.col(HARD_MINIMUM).is_null() | (value >= pl.col(HARD_MINIMUM))
+    below_maximum = pl.col(HARD_MAXIMUM).is_null() | (value <= pl.col(HARD_MAXIMUM))
+
+    df = df.join(bounds.lazy(), on=DataSchema.code_name, how="left")
+    df = df.filter(above_minimum & below_maximum)
+    return df.drop(HARD_MINIMUM, HARD_MAXIMUM)
+
+
+def event_time_filter(stage_cfg: DictConfig) -> pl.Expr | None:
```
Sllambias: See review comment 2

ehr2meds/meds_stages/aggregate_numeric_metadata.py
```diff
@@ -0,0 +1,216 @@
...
+        cutoff = datetime.strptime(str(configured_cutoff), "%Y-%m-%d")
+    except ValueError as error:
+        raise ValueError("event_time_cutoff must be a date in YYYY-MM-DD format") from error
+
+    return pl.col(DataSchema.time_name) < cutoff
+
+
+def fit_transform(
+    values: list[float],
+    *,
```
Sllambias: Is this necessary?

Sllambias (commented):
Overall I think the changes have improved the stages significantly.

I think the remaining changes would be to 
1. abstract the hardcoded string values out into the actual config files.
2. remove/reduce the passing of arbitrary dicts through functions and replace them with the arguments the function uses (e.g. `def read_value_bounds(configured_bounds) -> pl.DataFrame | None:` instead of `def read_value_bounds(stage_cfg: DictConfig) -> pl.DataFrame | None:`
3. Make sure we need the "*" star-expression flexibility and that it's not just an silent anti-crash feature

ehr2meds/meds_stages/aggregate_numeric_metadata.py
```diff
@@ -0,0 +1,216 @@
...
+
+    return {
+        LOWER_BOUND: lower,
+        UPPER_BOUND: upper,
+        BIN_EDGES: edges,
+        BIN_REPRESENTATIVES: representatives,
+    }
+
+
+def mapper_fntr(stage_cfg: DictConfig, code_modifiers: list[str] | None = None) -> Callable[[pl.LazyFrame], pl.LazyFrame]:
```
Sllambias, line 139: See review comment 2

ehr2meds/meds_stages/aggregate_numeric_metadata.py
```diff
@@ -0,0 +1,216 @@
...
+            max_bins=max_bins,
+            lower_quantile=lower_quantile,
+            upper_quantile=upper_quantile,
+        )
+        record.update(transform)
+        records.append(record)
+    return pl.DataFrame(records).sort(key)
+
+
+def reducer_fntr(stage_cfg: DictConfig, code_modifiers: list[str] | None = None) -> Callable[..., pl.DataFrame]:
```
Sllambias, line 214: See review comment 2

ehr2meds/meds_stages/annotate_numeric_values.py
```diff
@@ -0,0 +1,197 @@
...
+    LOWER_BOUND,
+    UPPER_BOUND,
+)
+from meds import CodeMetadataSchema, DataSchema
+from MEDS_transforms.stages import Stage
+from MEDS_transforms.utils import PKG_PFX, resolve_pkg_path
+from omegaconf import DictConfig
+from pathlib import Path
+
+NORMALIZED_VALUE = "numeric_value_normalized"
```
Sllambias: See review comment 1

ehr2meds/meds_stages/annotate_numeric_values.py
```diff
@@ -0,0 +1,197 @@
...
+            return pl.read_parquet(path)
+        case ".json":
+            return pl.read_json(path)
+        case _:
+            raise ValueError(
+                "numeric_metadata_filepath must be a JSON or Parquet file, or a directory containing codes.parquet"
+            )
+
+
+def prepare_metadata(metadata: pl.DataFrame, *, key: list[str], label: str = "numeric") -> pl.DataFrame:
```
Sllambias: See review comment 3

ehr2meds/meds_stages/annotate_numeric_values.py
```diff
@@ -0,0 +1,197 @@
...
+    fitted_metadata = prepare_metadata(fitted_metadata, key=key, label="fitted")
+
+    # External rows come first, so they take precedence for matching codes.
+    combined = pl.concat([external_metadata, fitted_metadata], how="vertical_relaxed")
+    combined = combined.unique(subset=key, keep="first", maintain_order=True)
+    return combined.sort(key)
+
+
+def is_usable(
+    *,
```
Sllambias: See review comment 3

ehr2meds/meds_stages/annotate_numeric_values.py
```diff
@@ -0,0 +1,197 @@
...
+    """Identify finite values that satisfy optional hard bounds."""
+    is_finite = value.is_not_null() & value.is_finite()
+    above_minimum = hard_minimum.is_null() | (value >= hard_minimum)
+    below_maximum = hard_maximum.is_null() | (value <= hard_maximum)
+    has_transform = lower_bound.is_not_null()
+    return is_finite & above_minimum & below_maximum & has_transform
+
+
+def normalize(*, value: pl.Expr, lower_bound: pl.Expr, upper_bound: pl.Expr) -> pl.Expr:
+    """Clip to fitted percentiles and normalize to the interval [0, 1]."""
```
Sllambias, line 102: See review comment 3

ehr2meds/meds_stages/annotate_numeric_values.py
```diff
@@ -0,0 +1,197 @@
...
+    clipped = value.clip(lower_bound, upper_bound)
+    normalized = (clipped - lower_bound) / (upper_bound - lower_bound)
+    return pl.when(upper_bound <= lower_bound).then(0.0).otherwise(normalized)
+
+
+def calculate_bin_index(*, normalized: pl.Expr, edges: pl.Expr) -> pl.Expr:
```
Sllambias: See review comment 3

ehr2meds/meds_stages/annotate_numeric_values.py
```diff
@@ -0,0 +1,197 @@
...
+    # The edge lists are small and bounded, so a row-local struct is clear and safe.
+    row = pl.struct(edges.alias(EDGES_FIELD), normalized.alias(NORMALIZED_FIELD))
+    return row.map_elements(
+        find_bin,
+        return_dtype=pl.Int32,
+    )
+
+
+def annotation_columns(
+    *,
```
Sllambias: See review comment 3

ehr2meds/meds_stages/annotate_numeric_values.py
```diff
@@ -0,0 +1,197 @@
...
+    representative = bin_representatives.list.get(bin_index, null_on_oob=True)
+
+    return [
+        pl.when(usable).then(normalized).cast(pl.Float32).alias(NORMALIZED_VALUE),
+        pl.when(usable).then(bin_index).cast(pl.Int32).alias(BIN_INDEX),
+        pl.when(usable).then(representative).cast(pl.Float32).alias(BINNED_VALUE),
+    ]
+
+
+def annotate_numeric_values(data: pl.LazyFrame, metadata: pl.DataFrame, *, key: list[str]) -> pl.LazyFrame:
```
Sllambias: See review comment 3

ehr2meds/meds_stages/annotate_numeric_values.py
```diff
@@ -0,0 +1,197 @@
...
+        bin_edges=pl.col(BIN_EDGES),
+        bin_representatives=pl.col(BIN_REPRESENTATIVES),
+    )
+    annotated = annotated.with_columns(columns)
+    return annotated.drop(*METADATA_COLUMNS)
+
+
+@Stage.register(output_schema_updates=DERIVED_COLUMNS)
+def annotate_numeric_values_fntr(
+    stage_cfg: DictConfig,
```
Sllambias, line 185: See review comment 2

ehr2meds/meds_stages/join_numeric_bins.py
```diff
@@ -0,0 +1,40 @@
...
+
+from __future__ import annotations
+
+import polars as pl
+from collections.abc import Callable
+from meds import DataSchema
+from MEDS_transforms.stages import Stage
+from omegaconf import DictConfig
+
+CODE = DataSchema.code_name
```
Sllambias: See review comment 1

ehr2meds/meds_stages/join_numeric_bins.py
```diff
@@ -0,0 +1,40 @@
...
+    joined_code = pl.concat_str(
+        pl.col(CODE),
+        pl.lit(separator),
+        pl.col(NUMERIC_BIN).cast(pl.String),
+    )
+
+    return data.with_columns(pl.when(pl.col(NUMERIC_BIN).is_not_null()).then(joined_code).otherwise(pl.col(CODE)).alias(CODE))
+
+
+def join_numeric_bins_fntr(stage_cfg: DictConfig) -> Callable[[pl.LazyFrame], pl.LazyFrame]:
```
Sllambias, line 30: See review comment 2

ehr2meds/meds_stages/join_numeric_bins.py
```diff
@@ -0,0 +1,40 @@
...
+from collections.abc import Callable
+from meds import DataSchema
+from MEDS_transforms.stages import Stage
+from omegaconf import DictConfig
+
+CODE = DataSchema.code_name
+NUMERIC_BIN = "numeric_value_bin"
+
+
+def join_numeric_bins(data: pl.LazyFrame, *, separator: str = "//") -> pl.LazyFrame:
```
Sllambias: See review comment 3

Sllambias (approved):
Super nice, I'm super satisfied with how this is implemented so i'm happy to approve @mikkelfo do you have anything else?

mikkelfo (approved):
I'm fine with it, I think Sebastians's comments helped clean up the things I was worried about

PR #63, opened 2026-07-02
Fix fetal config

Montgomeryyyy:
draft of fix for fetal data, but still having issues with the set time when adding the hours and minutes: `time: set_time($date_start as '%Y-%m-%d', f"{$V_OTIME}:{$V_OMINUT}" as '%H:%M')`

Screenshot 2026-07-02 at 09 37 23

@Sllambias  @mikkelfo 

## Summary by CodeRabbit

* **New Features**
  * Added expanded fetal synthetic-data scenarios covering diagnoses, procedures, laboratory results, medications, admissions, smoking, and alcohol information.
  * Added support for richer timestamps, including fractional seconds and microsecond precision.
  * Added new fetal conversion profiles and event formatting for end-to-end data processing.

* **Improvements**
  * Improved handling of mapped values, missing records, composed fields, and large input files.
  * Enabled parallel processing of data tables to improve conversion performance.

* **Documentation**
  * Updated synthetic-data generation and conversion instructions for the new workflow.

PR #68, opened 2026-07-22
Update fetal configs for numerics

Montgomeryyyy:
Updated fetal configs combining #63 and #67

## Summary by CodeRabbit

* **New Features**
  * Updated the end-to-end DeepFetal workflow (raw → preMEDS → MEDS) to use new DeepFetal synthetic configs and the updated MEDS event mapping.
  * Enhanced the MEDS pipeline to augment event conversion and add numeric lab/value metadata processing and binning stages.
  * Improved preMEDS conversion performance by running table processing in parallel based on the configured worker count.
* **Bug Fixes**
  * Value mapping now preserves non-mapped values.
  * Better duplicate-row reporting and improved CSV parsing behavior.
* **Documentation**
  * Updated README and synthetic-data instructions with the new DeepFetal command examples and paths.


Sllambias (changes requested):
(I will be looking into making these changes myself just noting them here)

Making these changes:
1. Fetal_synth_event should not be necessary, and if it is then it's because our synthetic data is insufficient I think, so I'm trying to merge `fetal_ngc_event` and `fetal_synth_event` to `fetal_event`
2. Making DeepFetal subfolders in the config directories so e.g. preMEDS/fetal_... all get moved into preMEDS/DeepFetal to unclutter for other projects.
3. Making tests pass without making separate configs for synthetic data. They must pass ON the synthetic data with the real preMEDS and MEDS configs.
4. Changing some fetal configurations on pathing to just use default hydra pathing, should be irrelevant for everyone else than Maria and I

configs/MEDS/default_pipeline.yaml
```diff
@@ -7,34 +7,57 @@ description: |-
...
+  - augment_event_config:
+      source_event_conversion_config_fp: ${source_event_conversion_config_fp}
+      output_event_conversion_config_fp: ${event_conversion_config_fp}
+      event_columns:
+        row_idx: $row_idx
+  - shard_events:
+      infer_schema_length: 999999999
+      data_input_dir: ${input_dir}/data
   - split_and_shard_subjects
+    #   split_fracs:
```
Sllambias: Should this be removed or un-commented?

configs/MEDS/default_pipeline.yaml
```diff
@@ -7,34 +7,57 @@ description: |-
...
+    #       train: 0.9
+    #       tuning: 0.1
   - convert_to_subject_sharded
-  - convert_to_MEDS_events
-  - merge_to_MEDS_cohort
+  - convert_to_MEDS_events:
+      do_dedup_text_and_numeric: true # If true, nullify ``text_value`` when it equals ``numeric_value``.
```
Sllambias, line 36: Is there a reason that we only do this for the _with values_ config? I'd think we should be consistent where possible

configs/MEDS/default_pipeline.yaml
Montgomeryyyy: I think for my use-case it should be removed as I use the load-in split file function, but I am unsure what we want our default pipeline to do.

configs/MEDS/default_pipeline.yaml
Montgomeryyyy: We don't have "numeric_value" in the non_values config, so I suppose it doesn't really do anything without.

configs/MEDS/default_pipeline.yaml
RH-MikkelWerling: I suspect that most people would be running out of time evaluation - we should probably have split for pretraining here as a default I would say. Don't know if it matters too much though.

ehr2meds/preMEDS/extractor.py
```diff
@@ -1,12 +1,49 @@
...
 from ehr2meds.preMEDS.data_handler import DataHandler
 from ehr2meds.preMEDS.processors import Processor
+from multiprocessing import Pool
 from tqdm import tqdm
 from typing import Dict, Optional, Union
 
 logger = logging.getLogger(__name__)
 
 
+def process_single_table_worker(args):
```
mikkelfo, line 11: Doesn't this function need to keep track of the `row_index_start` as the other one does?

ehr2meds/convert_raw_to_premeds.py
mikkelfo, line 1: Why was this changed?

mikkelfo (commented):
Two things:

1. A couple of the CodeRabbit comments seem valid to fix
2. I believe the `row_index_start` has been removed by the multiprocessing function.

ehr2meds/convert_raw_to_premeds.py
Sllambias: Because it was setting up logging on top of the logging that Hydra is already setting up and doing it in different places, and the files/directories it was creating are already being created by Hydra

ehr2meds/preMEDS/extractor.py
Sllambias: Yes, I missed that. This was originally implemented in #63 and then hand-ported here where the mistake occured

configs/synthetic_generation/fetal_SDS_SP_from_pop_part3.yaml
Montgomeryyyy: I know that I put this in, but there isn't really anything on ngc called labtest, since it represented the CPMI - Prøvesvar which we now have put in by its real name. So maybe it makes sense to delete this.

configs/synthetic_generation/fetal_SDS_SP_from_pop_part2.yaml
Montgomeryyyy: should be called "kernefamilie"

configs/synthetic_generation/fetal_SDS_SP_from_pop_part2.yaml
```diff
@@ -108,46 +108,7 @@ data: # Mangler DAR
...
-              args:
-                min_val: 0
-                max_val: 300
-            - type: choice
-              args:
-                options: ["Positive", "Negative", "Ikke påvist", "påvist", "0 Rhd pos"]
-          probabilities: [0.4, 0.4, 0.2]
-
+        
   LPR2_ADM:
```
Montgomeryyyy: is t_adm which is already implemented in part3

configs/synthetic_generation/fetal_SDS_SP_from_pop_part2.yaml
```diff
@@ -372,10 +349,10 @@ linked_data: # Data that needs to be linked to the data above
...
-      contact_id_new: 
+      dw_ek_kontakt: 
         type: choice
-        file: kontakter
-        linked_on: contact_id
+        file: kontakter_all
+        linked_on: dw_ek_kontakt
         
 
   LPR2_procedures:
```
Montgomeryyyy: is "t_sksube" which is in part3, so can be deleted here

configs/synthetic_generation/fetal_SDS_SP_from_pop_part2.yaml
Montgomeryyyy: is "t_diag" which is already in part 3

configs/synthetic_generation/fetal_SDS_SP_from_pop_part3.yaml
```diff
@@ -29,8 +29,8 @@ linked_data:
     linked_columns:
       CPR:
         type: choice
-        file: patient_info
-        linked_on: CPR
+        file: Population
+        linked_on: MOR_CPR
   
   diagnosis:
```
Montgomeryyyy: I think this file was Diagnoseliste from CPMI (SP), so can be deleted

configs/synthetic_generation/fetal_SDS_SP_from_pop_part3.yaml
Montgomeryyyy: Is CPMI ADT hændelser from SP, so can be deleted

configs/synthetic_generation/fetal_SDS_SP_from_pop_part2.yaml
```diff
@@ -580,50 +557,79 @@ linked_data: # Data that needs to be linked to the data above
         file: nyfoedte
         linked_on: DW_EK_Foedsel
 
-  patient_info:
+  Mor - CPMI - ADT hændelser:
```
Montgomeryyyy, line 479: missing: 

ADT_haendelse:
        type: choice
        args:
          options: ["Indlæggelse", "Flyt_ind", "Tilbage fra orlov"]

configs/synthetic_generation/fetal_SDS_SP_from_pop_part3.yaml
```diff
@@ -369,8 +465,8 @@ linked_data:
     linked_columns:
       CPR:
```
Montgomeryyyy: MOR_CPR

configs/synthetic_generation/fetal_SDS_SP_from_pop_part3.yaml
```diff
@@ -369,8 +465,8 @@ linked_data:
...
         type: choice
-        file: patient_info
-        linked_on: CPR
+        file: Population
+        linked_on: MOR_CPR
 
   procedure:
```
Montgomeryyyy: does not exist

Montgomeryyyy (commented):
I think things look quite reasonable:) 
I've left a bunch of very nitpicky comments on the synthetic data generation which I realise are also things I should have just fixed, but I only now noticed that we are generating a lot of duplicate data with the more realistic synthetic data and the older more abstract data.

configs/MEDS/DeepFetal/fetal_ngc_event.yaml
RH-MikkelWerling, line 1: The float trick was needed for me, but it might only relate to my data. If it works, it works - but you might be able to just convert to integers without having the intermediate float operation. 
Additionally; check in your MEDS data whether it is only GENDER which has NA in the time column. I had some NAs that didn't cause any errors but were caused by ill-formed format strings.

configs/MEDS/default_pipeline.yaml
```diff
@@ -7,34 +7,57 @@ description: |-
...
+  - merge_to_MEDS_cohort:
+      additional_sort_by: [row_idx]
   - extract_code_metadata
+  - aggregate_numeric_metadata:
```
RH-MikkelWerling, line 40: Just a note; much if this should already be specified in the default config for numeric values. If you like to be verbose, this will also work and be fine, but note that this can be left out; see: https://github.com/FGA-DIKU/ehr2meds/blob/update_fetal_configs/configs/MEDS/default_numeric_values.yaml

RH-MikkelWerling (approved):
Looks good - very few comments from me. A caveat is that I'm really out of the loop concerning what you are doing with the synthetic data, so I've more or less skipped those parts.

configs/MEDS/DeepFetal/fetal_ngc_event.yaml
RH-MikkelWerling: Note; I'm speaking specifically about: 

```
SDS_t_sksopr:
  SDS_t_sksopr:
    code: f"RPS//{$C_OPR}"
    time: set_time($D_ODTO as '%Y-%m-%d', f"{(($V_OTIME as float) as int)}:{(($V_OMINUT as float) as int)}" as '%H:%M')

SDS_t_sksube:
  SDS_t_sksube:
    code: f"RPO//{$C_OPR}"
    time: set_time($D_ODTO as '%Y-%m-%d', f"{(($V_OTIME as float) as int)}:{(($V_OMINUT as float) as int)}" as '%H:%M')
  procedure_type:
    code: f"RPO_TYPE//{coalesce($procedure_type, 'P')}"
    time: set_time($D_ODTO as '%Y-%m-%d', f"{(($V_OTIME as float) as int)}:{(($V_OMINUT as float) as int)}" as '%H:%M')
```

configs/MEDS/default_pipeline.yaml
Sllambias: No, it shouldn't be configured in two parallel places so I'd remove it from here.

configs/MEDS/default_pipeline.yaml
```diff
@@ -7,34 +7,57 @@ description: |-
...
+      lower_quantile: 0.01
+      upper_quantile: 0.99
+      min_bins: 2
+      max_bins: 100
+      # Optional hard plausibility limits. Values outside these ranges retain
+      # their raw numeric_value but receive null normalized/binned values.
+      numeric_value_bounds: {}
+      # Example:
+      #   "L//NPU01566": {min: 0, max: 250}
+      numeric_metadata_output_filepath: ${output_dir}/numeric_metadata.json
```
Sllambias, line 42: The numeric_metadata_output_filepath and numeric_metadata_filepath sound like they should be the same to me.

RH-MikkelWerling: Hmmm I disagree. The numeric_metadata_filepath is supposed to be the external metadata, which overwrites your own metadata. The file that you generate (numeric_metadata_output_filepath) is what can act as the external numeric_metadata_filepath for others.

Sllambias: Fair, it's just like "checkpoint_filepath" which I also find a bit ambiguous - is it where you write your checkpoints or the checkpoint that you would like to read and start from?

RH-MikkelWerling: yeah, good point. It would probably be a good addition to relabel "numeric_metadata_filepath" as "external_numeric_metadata_filepath" or something that's shorter but with the same semantic meaning. I agree that the naming is a bit confusing.

ehr2meds/preMEDS/extractor.py
```diff
@@ -64,40 +106,17 @@ def get_subject_id_mapping(self) -> Union[None, Dict[str, int]]:
...
+        # Prepare arguments for the worker pool
+        worker_tasks = [
+            (table_name, table_config, self.cfg.paths.output, self.cfg.write_file_type, self.cfg.chunksize, subject_id_mapping)
+            for table_name, table_config in self.cfg["tables"].items()
+        ]
 
-            self.data_handler.save(processed_chunk, table_name)
+        logger.info(f"Starting multiprocessing pool with {self.cfg.num_workers} workers.")
 
-            if self.cfg.get("test"):
```
Montgomeryyyy, line 102: we seem to have removed the test functionality?

Sllambias: It seemed like a somewhat unfinished functionality given that it's just an purposely broken for-loop, but if it's functionality that was very convenient I can make sure we get something similar. 
What was the purpose of it? To produce a mini-version of the real output file?

configs/MEDS/DeepFetal/fetal_ngc_event.yaml
Sllambias: @Montgomeryyyy Do you have anything to note here? Otherwise I'll mark it as resolved :)
At least it runs on NGC, but I have not looked in depth at the produced files

Sllambias (approved):
@Montgomeryyyy Once we have solved the last two comments I think we're ready to merge this and close #63 
@mikkelfo if you would like to also review please let me know :)

ehr2meds/preMEDS/extractor.py
Montgomeryyyy: Yeah I've used it a lot a least, to just have a subset version of the real output file for testing things both in ehr2meds and bonsai, where you still have the whole range of data sources but just less data.

Sllambias: Okay fair, I'll make sure we keep the old or get something similar before I merge

Sllambias (changes requested):
Test fn

configs/MEDS/DeepFetal/fetal_ngc_event.yaml
Montgomeryyyy: I've looked at the files, and they output usable timestamps, so I would say it works, but I'm not sure I ever tried without. For the NA timestamps we don't seem to have any as we have "." instead, which I have tried to replace with null values in the preMEDS stage.

ehr2meds/preMEDS/extractor.py
mikkelfo: Yeah I removed it in a previous PR but got multiple requests to add it back, so we definitely need some way of just checking the entire pipeline without processing all chunks

mikkelfo (approved):
Mainly touches a bunch of .yaml files that is deep fetal specific, so I don't have any additional comments

Sllambias (approved):
reintroduced test so all comments should be addressed

